PR #35, opened 2015-07-22
In _increasePoolIfNeeded, check that physicalSize != optPhysicalSize

Brian-K-Smith:
If the two are equal, missingItems = 0 / this._physicalAverage = NaN and the nextPhysicalCount = NaN and delta = NaN and we pass NaN to this._createPool and new Array, which throws errors.


Brian-K-Smith:
I signed it!

Brian-K-Smith:
Apparently this fix breaks a bunch of other browsers, will keep looking at it tomorrow.

PR #113, opened 2015-10-02
resize -> iron-resize to prevent more head banging

danbeam:
The event listened for is actually `iron-resize`, not `resize`.


frankiefu:
LGTM

PR #174, opened 2016-01-13
Automatic cleanup!

rictic:


rictic:
Tedium's markdown to markdown transpiler does em / strong backwards. Gotta fix that.

PR #337, opened 2016-10-26
Fix case where obsolete cached value _viewportHeight is used.

freshp86:
Fix case where obsolete cached value `_viewportHeight` is used, causing the iron-list to only show 3 items, even though the container has enough height to fit more items.

### DIFF
--- a/iron-list.html
+++ b/iron-list.html
@@ -1390,11 +1390,14 @@ will only render 20.
 
       forceUpdate = forceUpdate || this._scrollHeight === 0;
       forceUpdate = forceUpdate || this._scrollPosition >= this._estScrollHeight - this._physicalSize;
+      // Probably a bug that style.height is a string like '100px' and it is being compared to
+      // a number?
       forceUpdate = forceUpdate || this.grid && this.$.items.style.height < this._estScrollHeight;
 
       // Amortize height adjustment, so it won't trigger large repaints too often.
       if (forceUpdate || Math.abs(this._estScrollHeight - this._scrollHeight) >= this._optPhysicalSize) {
         this.$.items.style.height = this._estScrollHeight + 'px';
+        this.updateViewportBoundaries();
         this._scrollHeight = this._estScrollHeight;
       }
     },


freshp86:
/cc @blasten @danbeam 

Several past reports from users about iron-list incorrectly populated with 3 items only on startup have been dismissed as "need to properly size the iron-list's parent", but I think there is a bug inside iron-list where it uses an obsolete `_viewportHeight` value causing it to not increase the physicalSize even though there is enough height.

This PR updates the cached values right after the container of the iron-list has been assigned a new height, so that follow up call to `this._increasePoolIfNeeded()` from within `scrollToIndex()`  uses the updated value, and fixes the issue.

iron-list.html
```diff
@@ -1390,11 +1390,14 @@
 
       forceUpdate = forceUpdate || this._scrollHeight === 0;
       forceUpdate = forceUpdate || this._scrollPosition >= this._estScrollHeight - this._physicalSize;
+      // Probably a bug that style.height is a string like '100px' and it is being compared to
```
blasten: good finding

danbeam:
any updates on this?  /cc @tjsavage

iron-list.html
```diff
@@ -1390,11 +1390,14 @@
...
+      // a number?
       forceUpdate = forceUpdate || this.grid && this.$.items.style.height < this._estScrollHeight;
 
       // Amortize height adjustment, so it won't trigger large repaints too often.
       if (forceUpdate || Math.abs(this._estScrollHeight - this._scrollHeight) >= this._optPhysicalSize) {
         this.$.items.style.height = this._estScrollHeight + 'px';
+        this.updateViewportBoundaries();
```
blasten: This change would force a style-recalc.

freshp86: Is there a better way to make the code not use obsolete values in follow up calculations? I mean, yes, this will cause a style-recalc, but it also fixes the calculations to be correct passed that point.

blasten: Do you have a repro case? My fear is that there's some async element upgrade happening that affects the layout after the list started rendering. See [this line](https://github.com/PolymerElements/iron-list/blob/856ebcf90b2834aae728fc295b5dd5aa5ebc18ca/iron-list.html#L1004).

freshp86: I have a repro case within the Chromium codebase new MD Settings page. Have not spent time to create a minimal repro case yet, but I'll do if you think this is useful.

In short, it happens when we render an iron-list while it is not visible (on purpose). Then the user navigates to the section that causes the iron-list to become visible and it only has 3 elements. Let me know if you think the time for a minimal repro is worth it.

blasten: @freshp86 We have a behavior called `IronResizableBehavior` which provides the resize/visibility signal to elements like iron-list. Would you mind telling me how to repro the issue in the MD Settings page?

Westbrook: As a person who's told a large number of people "need to properly size the iron-list's parent" myself, I'd personally love to see a repro where that actively wasn't the case, if for nothing else than to add some empathy to my response next time. The text in your "in short" makes me think there's a `this.notifyResize()` missing, but I'd love to know better when making suggestions to others on that point, too!

danbeam: @blasten https://bugs.chromium.org/p/chromium/issues/detail?id=657336#c12

Westbrook: @danbeam Looking through this example, I don't see anything that would let the list (at least cr-network-list) know that is had come into view. Maybe there's something at the page level, further up than I chose to investigate, but if you're really placing the list on the DOM while invisible and then making it visible, you have to wait till it is visible for `this.notifyResize()` to be able to actually do its job. In Crome Settings' case, this might be a perfect opportunity to employ an Intersection Observer, though that may just be me wanting a good reason to use one. I do see how the one forced recalc in this PR is better than the approach in the issue you linked, being it would be doing a lot more mathing of the page (seemingly over and over again in empty lists) but as they say "a stitch in time saves nine".

freshp86: Unfortunately the previous repro example in MD Settings has been affected a bit by a follow up CL, so the repro instructions are pretty involved (reverting previous commits, rebuilding Chromium).

I tried to create a minimal reproduction case at http://jsfiddle.net/s6f029j3/12/, which I think is similar to the MD Settings one. I don't understand, why even though iron-list figures out the correct height (see red rectangle in the example), it refuses to render more elements until resize. From reading previous discussions on the matter. Why is the default behavior so surprising (as revealed by numerous discussions on iron-list sizing)? Are you sure that asking devs to call anything after the list is rendered, is the right approach API-wise? Asking especially since this patch seems to fix the issue.

blasten: @freshp86 the selector that sets the size should be inside the shadow root. http://jsfiddle.net/wta2qr88/1/

freshp86: In my example there is no CSS rule to specify the size, on purpose. I would understand your point better, if resizing the window had absolutely no effect when I don't have an explicit size in CSS (keep showing 3 elements only).

The fact that resizing the window results in more elements shown, even though no size is specified in CSS makes it very confusing. Which brings me to my question, is iron-list using obsolete cached values on purpose? Is this CL wrong from a logic perspective, or is it just making things more expensive?

Westbrook: Your example is interesting in that with Shadow DOM on there is no reason why the list would ever be 200px tall. 

To correct for that, you don't need to call anything as long as the following instructions are accurately fulfilled:

> Sizing iron-list
> 
> iron-list must either be explicitly sized, or delegate scrolling to an explicitly sized parent. By "explicitly sized", we mean it either has an explicit CSS height property set via a class or inline style, or else is sized by other layout means (e.g. the flex or fit classes).

Checkout an altered version of your sample here: http://jsfiddle.net/vLv0fdk8/ I've set the height in exact pixels as per your sample, but you could equally use `display: flexbox` or `postion: absolute` with top and bottom properties set, or some others that I can't think of right now for the same results.

blasten: @freshp86 In your example, the list's scrollTarget (#container) has `offsetHeight = 0`.  

The iron-list's `_render` runs with that state, it renders the initial count which is `3`. Then based on those 3 items, the content size is calculated and set by `_updateScrollerSize`. However, because `#container` has no size, setting the size on a descendant will also affect the size of `#container`. When you resize the browser, the list takes the new scroll target's `height` which is incorrect anyways. 

If we were updating the height as you proposed, the list will keep rendering nodes until the limit is reached. To see that, change your loop to: ` for (var i = 0; i < 600; i++)` and resize the browser window.

Whatever change we make will not be correct, and I do agree that it would be great to throw or indicate in the console that something is wrong. I just haven't found a good way to do the detection since in many cases, the list is in a container with `display=none`.

Westbrook: But if the list or its container don't have an explicit height then this happens: http://jsfiddle.net/L54tyyr5/ 

Notice *ALL* of the DOM nodes...

blasten: Also, looking at the stack trace of the Settings page:

screen shot 2016-11-01 at 10 09 12 am

It's calling `notifyResize()` from a `setInterval`.  Another way is to set fire `notifyResize` only when the animated page first renders (before the animation starts). However that would require that the scroll target's height is set. See: https://github.com/PolymerElements/iron-list#sizing-iron-list

freshp86: > However, because #container has no size, setting the size on a descendant will also affect the size of #container. When you resize the browser, the list takes the new scroll target's offsetHeight which is incorrect anyways.

Why are you considering the new scroll target's `offsetHeight` incorrect? Or let me rephrase the question. Can an `iron-list` not be used as a "dumb" list instead of a "smart" list, such that all elements are always rendered, unless there is an explicit size set? This would make the default behavior more intuitive and it would match closer what the browser does in similar cases (as you stated in your comment above).

blasten:
Closing in favor of https://github.com/PolymerElements/iron-list/pull/345

PR #379, opened 2017-01-19
child templates don't update when items updates

### LINKED ISSUE #300: iron-list not maintaining bindings in items array

admwx7:
When changing the array of items, the child templates should get updated, however the dirty checking doesn't trigger updates when setting an object to an existing object, so to get around the dirty checking the object needs to be cleared first.

fixes #300
@blasten

### DIFF
--- a/iron-list.html
+++ b/iron-list.html
@@ -1254,6 +1254,7 @@ will only render 20.
         var inst = el._templateInstance;
         var item = this.items && this.items[vidx];
         if (item != null) {
+          inst[this.as] = {};
           inst[this.as] = item;
           inst.__key__ = this._collection.getKey(item);
           inst[this.selectedAs] = /** @type {!ArraySelectorElement} */ (this.$.selector).isSelected(item);


iron-list.html
```diff
@@ -1254,6 +1254,7 @@
         var inst = el._templateInstance;
         var item = this.items && this.items[vidx];
         if (item != null) {
+          inst[this.as] = {};
```
blasten: I see. However, this will make bindings run with `undefined` values. see this thread: https://github.com/PolymerElements/iron-list/pull/328#discussion_r82298950

admwx7:
This is actually the recommended solution in the documentation for handling
updates that override the dirty checks, when using es6 you can just
Object.assign to get around the issue.

PR #427, opened 2017-04-08
add Method getPhysicalItemForItem

jogibear9988:
This Method is usefull for example, if you want to access the created dom node, after selecting an Item. (I use it to call setfocus on an element)

### DIFF
--- a/iron-list.html
+++ b/iron-list.html
@@ -1461,6 +1461,17 @@ will only render 20.
     },
 
     /**
+     * Get's the created Physical Item for a given Item (if it exists).
+     *
+     * @method getPhysicalItemForItem
+     * @param {Object} item The item instance.
+     */
+    getPhysicalItemForItem: function(item) {
+        let index = this.items.indexOf(item);
+        return this._physicalItems[this._getPhysicalIndex(index)];
+    },
+    
+    /**
      * Selects the given item.
      *
      * @method selectItem


jogibear9988:
I signed it!

keanulee:
A physical item may not always exist for a given item, since the list is virtualized. Since the creation/usage of physical items is handled by iron-list, it should be considered as a private implementation detail.

jogibear9988:
I know, but I need to access the physical item when it exists... I want to set focus to a textbox when a item is selected!

jogibear9988:
Any news if this could be added?

keanulee:
We're hesitant on adding this method for the reason mentioned above, but we're considering adding a `focusItem(virtualIndex)` method that is essentially an alias of the private `_focusPhysicalItem` method (https://github.com/PolymerElements/iron-list/compare/blur-removed-item). The advantage of this is that this would scroll the item into view (if it isn't already). If you need to further access the element, you can use `document.activeElement`. WDYT of this idea?

jogibear9988:
Ok!

One more thing, for example in Wpf, where you have a VirtualizingStackPanel, there are also methods, to get the PhysicalItem from the data, and if it does not exist you get null

jogibear9988:
Or we create events (or callbacks) physical-item-created and physical-item-removed wich have the dataitem and the physical item as a parameter

PR #430, opened 2017-04-24
Grammar tweaks.

davidmaxwaterman:
The only 'error' was in the phrase 'bring in a page at *the* time', but I changed it to read better (imo).


davidmaxwaterman:
Actually, I just noticed I missed out a word :( `it is better *to* split the data into pages`. Sigh.

PR #433, opened 2017-04-29
Add mapping for left/right and grid specific arrow key interactions

### LINKED ISSUE #432: Grid based keyboard navigations
### Description
When the `grid` attribute is present, the keyboard navigation offered by the up and down arrow binding fall short of the short of that specified by WAI-ARIA for [Layout Grids](https://www.w3.org/TR/wai-aria-practices-1.1/#h-layoutgrid).

### Expected outcome
When the `grid` attribute is present, the following keys provide grid navigation by moving focus among cells of the grid. These key commands are available by default after an element in the grid receives focus.

- Right Arrow: Moves focus one cell to the right. Optionally, if focus is on the right-most cell in the row, focus may move to the first cell in the following row. If focus is on the last cell in the grid, focus does not move.
- Left Arrow: Moves focus one cell to the left. Optionally, if focus is on the left-most cell in the row, focus may move to the last cell in the previous row. If focus is on the first cell in the grid, focus does not move.
- Down Arrow: Moves focus one cell down. Optionally, if focus is on the bottom cell in the column, focus may move to the top cell in the following column. If focus is on the last cell in the grid, focus does not move.
- Up Arrow: Moves focus one cell up. Optionally, if focus is on the top cell in the column, focus may move to the bottom cell in the previous column. If focus is on the first cell in the grid, focus does not move.

### Actual outcome
When the `grid` attribute is present, the following keys provide grid navigation by moving focus among cells of the grid. These key commands are available by default after an element in the grid receives focus.

- Down Arrow: Moves focus one cell to the right. Optionally, if focus is on the right-most cell in the row, focus may move to the first cell in the following row. If focus is on the last cell in the grid, focus does not move.
- Up Arrow: Moves focus one cell to the left. Optionally, if focus is on the left-most cell in the row, focus may move to the last cell in the previous row. If focus is on the first cell in the grid, focus does not move.

### Live Demo
https://user-content-dot-custom-elements.appspot.com/PolymerElements/iron-list/v1.4.4/iron-list/demo/grid.html

### Steps to reproduce
1. Put an `iron-list` element in the page with the `grid` attribute applied
2. Open the page in a web browser.
3. Focus the list.
4. Use the up and down arrows to move the focus around the list
5. Use the left and right arrows, which currently don't move the focus

### Browsers Affected

- [x] All

Westbrook:
Fixes #432

The included code extends the keyboard navigation interface to include left and right arrows when the `grid` attribute is present, and alters the response to the up and down arrows in accordance with the WAI-ARIA spec for [Layout Grids](https://www.w3.org/TR/wai-aria-practices-1.1/#h-layoutgrid).

Tests to cover this behavior is Grids, RTL Grids, and standard Lists, as well as slightly update README.md documentation is also included.

### DIFF
--- a/iron-list.html
+++ b/iron-list.html
@@ -1878,16 +1878,10 @@ will only render 20.
       this._focusPhysicalItem(this._focusedIndex - (this.grid ? this._itemsPerRow : 1));
     },
     _didMoveForward: function(e) {
-      // When `grid === true` the `upArrow` goes backward `this._itemsPerRow` items
-      var destination = this._focusedIndex + (this.grid ? this._itemsPerRow : 1);
-	  // When `grid === true` move focus down one row, either
-	  // directly below or the last item of the list.
-	  if (this.grid && this._focusedIndex < (this.items.length - (this.items.length % this._itemsPerRow))) {
-		  destination = Math.min(destination, this.items.length - 1);
-	  }
       // disable scroll when pressing the down key
       e.detail.keyboardEvent.preventDefault();
-      this._focusPhysicalItem(destination);
+      // When `grid === true` the `upArrow` goes backward `this._itemsPerRow` items
+      this._focusPhysicalItem(this._focusedIndex + (this.grid ? this._itemsPerRow : 1));
     },
     // When in grid mode, the use of `leftArrow` changes the focus based on the text direction
     _didMoveLeft: function(e) {
--- a/iron-list.html
+++ b/iron-list.html
@@ -1873,14 +1873,12 @@ will only render 20.
       }
     },
     // When `grid === true` the `downArrow` goes forward `this._itemsPerRow` items
-    _didMoveBackward: function(e, step) {
-      step = step || (!this.grid) ? 1 : this._itemsPerRow;
-      this._focusPhysicalItem(this._focusedIndex - step);
+    _didMoveBackward: function(e) {;
+      this._focusPhysicalItem(this._focusedIndex - (!this.grid ? 1 : this._itemsPerRow));
     },
     // When `grid === true` the `upArrow` goes backward `this._itemsPerRow` items
-    _didMoveForward: function(e, step) {
-      step = step || (!this.grid) ? 1 : this._itemsPerRow;
-      var destination = this._focusedIndex + step;
+    _didMoveForward: function(e) {;
+      var destination = this._focusedIndex + (!this.grid ? 1 : this._itemsPerRow);
 	  // When `grid === true` move focus down one row, either
 	  // directly below or the last item of the list.
 	  if (this.grid && this._focusedIndex < (this.items.length - (this.items.length % this._itemsPerRow))) {
@@ -1893,14 +1891,12 @@ will only render 20.
     // When in grid mode, the use of `leftArrow` changes the focus based on the text direction
     _didMoveLeft: function(e) {
       if (!this.grid) return;
-      if (this._isRTL) this._didMoveForward(e, 1);
-      else this._didMoveBackward(e, 1);
+      this._focusPhysicalItem(this._focusedIndex + (this._isRTL ? 1 : -1));
     },
     // When in grid mode, the use of `rightArrow` changes the focus based on the text direction
     _didMoveRight: function(e) {
       if (!this.grid) return;
-      if (this._isRTL) this._didMoveBackward(e, 1);
-      else this._didMoveForward(e, 1);
+      this._focusPhysicalItem(this._focusedIndex + (this._isRTL ? -1 : 1));
     },
 
     _didEnter: function(e) {
--- a/iron-list.html
+++ b/iron-list.html
@@ -1872,20 +1872,20 @@ will only render 20.
         }
       }
     },
-    // When `grid === true` the `downArrow` goes forward `this._itemsPerRow` items
-    _didMoveBackward: function(e) {;
-      this._focusPhysicalItem(this._focusedIndex - (!this.grid ? 1 : this._itemsPerRow));
+    _didMoveBackward: function() {
+      // When `grid === true` the `downArrow` goes forward `this._itemsPerRow` items
+      this._focusPhysicalItem(this._focusedIndex - (this.grid ? this._itemsPerRow : 1));
     },
-    // When `grid === true` the `upArrow` goes backward `this._itemsPerRow` items
-    _didMoveForward: function(e) {;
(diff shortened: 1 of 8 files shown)


Westbrook:
If I'm reading [this](https://travis-ci.org/PolymerElements/iron-list/builds/227154668) correctly, two of the tests I added failed in Chrome 58 when run with Travis and Sauce Labs. Currently, all tests are passing for me locally in Chrome 57/58/60, so I'm wondering if anyone has seen issues with Travis and Sauce Labs in conjunction with focus and/or keyboard input tests?

If there happened to be some flakiness in this area with Travis and Sauce, is it possible to somehow manually rerun these tests? Or, if there is something code specific that I can do to my test to be more resilient, I'd be happy to hear any ideas.

cc: @blasten

iron-list.html
```diff
@@ -1870,15 +1872,35 @@
...
       // disable scroll when pressing the down key
-      e.detail.keyboardEvent.preventDefault();
-      this._focusPhysicalItem(this._focusedIndex + 1);
+      if (e.detail.keyboardEvent.keyCode == 40) e.detail.keyboardEvent.preventDefault();
+      this._focusPhysicalItem(destination);
+    },
+    // When in grid mode, the use of `leftArrow` changes the focus based on the text direction
+    _didMoveLeft: function(e) {
+      if (!this.grid) return;
+      if (this._isRTL) this._didMoveForward(e, 1);
```
keanulee: Instead of calling another event handler (`_didMoveBackward`), just inline what it does here (and avoid overloading/remove the second param to `_didMoveBackward`):

```js
this._focusPhysicalItem(this._focusedIndex + (this._isRTL ? 1 : -1));
```

Westbrook:
@keanulee good call on the simplification.

I had hoped maybe that update to the test would put this over the top, alas. Any thoughts on getting that test to pass?

iron-list.html
```diff
@@ -1870,15 +1872,31 @@
...
-
-    _didMoveDown: function(e) {
+    // When `grid === true` the `upArrow` goes backward `this._itemsPerRow` items
+    _didMoveForward: function(e) {;
+      var destination = this._focusedIndex + (!this.grid ? 1 : this._itemsPerRow);
+	  // When `grid === true` move focus down one row, either
+	  // directly below or the last item of the list.
+	  if (this.grid && this._focusedIndex < (this.items.length - (this.items.length % this._itemsPerRow))) {
+		  destination = Math.min(destination, this.items.length - 1);
+	  }
```
keanulee: This if statement can be eliminated if you set the correct destination in one statement:

```js
var destination = Math.min(this._focusedIndex + (this.grid ? this._itemsPerRow : 1), this._virtualCount - 1);
```

or


```js
var destination = this._focusedIndex + (this.grid ? this._itemsPerRow : 1);
destination = Math.min(destination, this._virtualCount - 1);
```

iron-list.html
```diff
@@ -1870,15 +1872,31 @@
...
       // disable scroll when pressing the down key
-      e.detail.keyboardEvent.preventDefault();
-      this._focusPhysicalItem(this._focusedIndex + 1);
+      if (e.detail.keyboardEvent.keyCode == 40) e.detail.keyboardEvent.preventDefault();
```
keanulee: Don't need this conditional anymore (handler now only called for down arrow).

test/focus.html
```diff
@@ -104,13 +105,50 @@
...
+          MockInteractions.pressAndReleaseKeyOn(list, 38); // up
+        });
+        itemToFocus.addEventListener('focus', function(e) {
+          var focusedIndex = list.modelForElement(e.target).index;
+          assert.equal(focusedIndex, 0);
+          done();
+        });
+        MockInteractions.focus(initialItem);
+      });
+    });
```
keanulee: `focus` events don't fire when the test runner runs and the window is not focused. This can be replicated in Chrome if refresh the tab and click outside the window before the tests start running.

iron-list.html
Westbrook: The reasoning for the `if` statement is to allow the following set of elements (E) with focused item (F):
```
E E E E
E E F E
E E
```
To become the following when the DOWN arrow is pressed:
```
E E E E
E E E E
E F
```
But prevent the following:
```
E E E E
E E E E
F E
```
From becoming:
```
E E E E
E E E E
E F
```
Is there a better comment that I could provide?

test/focus.html
Westbrook: Weird that the test runner isn't focused on these two tests, but IS focused on that tests added below. Should I remove these or rewrite them in some way? Or?

Westbrook: @keanulee, updated the focus tests to rely on `flush()` and `element.tabIndex` as in some of the other tests in `focus.html`, so I'm finally getting greens on the remote testing pass!!

iron-list.html
Westbrook: I've added one more test to outline the difference between the code that is there and the change you are asking for here. I could see it going both ways, and am mainly working with the context laid out by the UX designer here at my office. If you felt the other way was more conducive to the broader functionality of `iron-list`, I can make this change, as well as add a corollary to the UP arrow interaction. Let me know what your thoughts are on this.

keanulee: I don't have a strong preference for one behavior over the other, but if this is the desired result, you can just call `_focusPhysicalItem(this._focusedIndex + (this.grid ? this._itemsPerRow : 1))` with an out of range index and let the first `if` statement of `_focusPhysicalItem` do the early return. That way, this method would mirror the `_didMoveBackward` handler. Keep the test you wrote though to make sure this behavior doesn't unintentionally change in the future.

Westbrook: It's unclear to me how the use of `_focusPhysicalItem(this._focusedIndex + (this.grid ? this._itemsPerRow : 1))` allows for both of the cases that I've outlined above to be true. My current goal is for `_didMoveForward` to NOT work exactly like `_didMoveBackward` in that moving forward can intersect with an incomplete row while moving backwards cannot.

If there is some concept of code simplicity or API contact requirements that prevent that from being true, I can completely understand. Just let me know that's the case, and my UX team and I can agree on a simpler, unified mechanism of moving forward and backward to submit to this PR.

keanulee: Ah, I see what you mean, though I prefer simplifying it so pressing the down arrow when there isn't an item directly below is a noop.

Westbrook: Understood. Updates pushed.

keanulee:
Thanks for the contribution @Westbrook

Issue #389, opened 2017-01-27
The iron-list jumps back down while scrolling up after having removed a few items.


thibaultzanini:
### Description
The `iron-list` jumps back down while scrolling up after having removed a few items.

### Expected outcome

The `iron-list` should scroll smoothly after removing items.

### Actual outcome

The `iron-list` jumps back while scrolling after removing items.

### Live Demo
http://jsbin.com/dusejihuho/1/edit?html,console,output

### Steps to reproduce

1. Scroll down the `iron-list` to the latest item.
2. Click the `remove 16 first items` button. 
3. Scroll up slowly.

http://d.pr/i/L01i

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Safari 10

PR #435, opened 2017-04-29
#434 correct relationship between _virtualStart and _physicalTop

### LINKED ISSUE #434: The relationship between `_virtualStart` and `_physicalTop` is not maintained
### Description
When "random access" scrolling occurs, the relationship between `_virtualStart` and `_physicalTop` is not maintained in such a way to ensure that the appropriate items are visible. This is particularly apparent when items are removed from the list and then a "random access" scroll occurs, see #389.

### Expected outcome
- Taking out items and then moving to a specific scrollTop should display the correct items.
- When starting at the end of a list of items with `height: 100px`, removing a series of elements and then executing `list.scrollTo(0,100)` should display you the item with index of 1, not 5, which invalidates the test found here: https://github.com/PolymerElements/iron-list/blob/master/test/mutations.html#L197

### Actual outcome
- When scrolled to the bottom of a standard list and items from the beginning of the list are removed, `list.scrollTo(0, {place close to the top})` should display the correct element.
- When the `grid` attribute is present and the last row contains a single item, shifting the list and subsequently moving the scroll to the end of the list should show the last element.

### Live Demo
https://jsbin.com/virovetuqo/1/edit?html,console,output

### Steps to reproduce
1. Put an `iron-list` element in the page.
2. Set it up to have three rows.
3. Give it 3n + 1 items.
4. Open the page in a web browser.
5. Remove the first item from the items array.
6. Programmatically change scrollTop so you are at the end of the list.
7. Observe that the last item is not currently displayed.

### Browsers Affected

- [x] All

Westbrook:
Fixes #434 

Acquires relationship approach from `scrollToIndex` rather than an static relationship to the scroll positions. This allows for `_physicalTop` to take into account that at "scrollTop === 0" it should be the same at the top of the visible space, and at "scrollTop === maxScrollTop" it should be offset by ~0.25 pages.

I've updated the test for #389 that appeared erroneous, and added one to address the context that you have one items on the last row of a grid and after removing the last item setting the scrollTop to the last row actual set you on a row of n - 2 as seen in the JS Bin in the issue.

### DIFF
--- a/test/grid.html
+++ b/test/grid.html
@@ -152,18 +152,18 @@ subject to an additional IP rights grant found at http://polymer.github.io/PATEN
     });
 
     test('delete a grid item when the last row should only have one item and scroll to bottom', function(done) {
-	  list.items = buildDataSet(64);
-	  Polymer.dom.flush();
-	  list.shift('items');
-	  Polymer.dom.flush();
-	  list.scroll(0, 10000000);
-	  requestAnimationFrame(function() {
-		setTimeout(function() {
-		  assert.equal(list.lastVisibleIndex, list.items.length - 1);
-		  done();
-		});
-	  });
-	});
+      list.items = buildDataSet(64);
+      Polymer.dom.flush();
+      list.shift('items');
+      Polymer.dom.flush();
+      list.scroll(0, 10000000);
+      requestAnimationFrame(function() {
+        setTimeout(function() {
+          assert.equal(list.lastVisibleIndex, list.items.length - 1);
+          done();
+        });
+      });
+    });
 
     test('Columns per row (#381)', function(done) {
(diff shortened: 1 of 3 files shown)


test/grid.html
```diff
@@ -151,6 +151,20 @@
       });
     });
 
+    test('delete a grid item when the last row should only have one item and scroll to bottom', function(done) {
+	  list.items = buildDataSet(64);
```
keanulee: nit: indentation seems off - should use 2 spaces (no tabs).

Westbrook: Whoops. Sorry about that.

keanulee:
Thanks @Westbrook

PR #437, opened 2017-04-30
Add selectItems and associated tests

### LINKED ISSUE #436: Select processing of large groups of selects takes too long
### Description
Not sure if this is ultimately an `iron-list`, or an `array-selector`, or a further up, issue, but selecting large groups of content, whether by the technique outlined in #124 or in #361 takes an inordinate amount of time. Explicitly, I'm seeing ~10s processing times for lists of 5000 items. I understand that's a lot of items, BUT that still seems crazy long.

My research points to this coming from the way [`selectItem` applied code from `array-selector`](https://github.com/PolymerElements/iron-list/blob/master/iron-list.html#L1588) When you follow that into the `array-selector`, you can see that it's [pushing into it's `selected` array](https://github.com/Polymer/polymer/blob/1.x/src/lib/template/array-selector.html#L201) the subsequent property effects of which seems to be the cause of the processing delay.

### Expected outcome
Selecting a large amount of items takes an virtually indistinguishably longer time than selecting one or a few.

### Actual outcome
Selecting 5000 items at a time causes a ~10s block on the main thread.

### Live Demo
https://jsbin.com/tobosaleki/1/edit?html,console,output

### Steps to reproduce
1. Visit the JS Bin demo
2. Click 'select all'
3. Wait, see the window.performance based timing display on completion
4. Click 'select none'
5. Wait, a really small amount of time, see the window.performance based timing display on completion
6. Be in awe at the difference

### Browsers Affected

- [x] All

Westbrook:
Fixes #436

As mentioned in the issue above, this may actually be an `array-selector` issue more than an `iron-list` issue, but as this specific code seems better managed in `iron-list` than in user space code as initially noted with the potential solution in #124, I figured this was a good step in getting this conversation more definitively completed.

This code adds a `selectItems` method that selects the entirety of `this.items` when no arguments are found, or casts the content of arguments into an Array such that they can be selected in batch regardless of them being an array, a single item, or a list of items.

Tests for "select items: all", "select items: array" (sending the method an array), "select items: arguments" (sending the batch as an arguments list), and "select items: single" to maintain this functionality moving forward.

### DIFF
--- a/iron-list.html
+++ b/iron-list.html
@@ -1623,7 +1623,7 @@ will only render 20.
       var key = '';
       // Cast an arguments list that doesn't consist of a single array to an array
       if (arguments.length > 1 || (typeof items !== 'undefined' && !Array.isArray(items))) {
-        items = Array.from(arguments);
+        items = !!Array.from ? Array.from(arguments) : [].slice.call(arguments);
       }
       items = items || this.items;
       items = items.map(function(item) {
--- a/iron-list.html
+++ b/iron-list.html
@@ -1628,11 +1628,11 @@ will only render 20.
       items = items || this.items;
       items = items.map(function(item) {
         return this._getNormalizedItem(item);
-      }.bind(this))
-      .filter(function(item) {
+      }.bind(this));
+      this.$.selector.set('selected', items);
+      items = items.filter(function(item) {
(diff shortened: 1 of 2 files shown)


keanulee:
Closing in favor of #457

PR #452, opened 2017-06-17
Update `lastVisibleIndex` and `firstVisibleIndex` when list is resized.

### LINKED ISSUE #451: `lastVisibleIndex` and `firstVisibleIndex` aren't updated correctly when the list is resized
### Description
When resizing an `iron-list`, the `lastVisibleIndex` and `firstVisibleIndex` can fall out of sync with the actual delivery of the component.

### Expected outcome
`lastVisibleIndex` and `firstVisibleIndex` should be as closely aligned to the actual delivery of the component as possible.

### Actual outcome
When resizing an `iron-list` that is not scrolled all the way to the bottom, the `lastVisibleIndex` should update when the list is resized. When scrolled at or close to the end of the list, resizing should update the `firstVisibleIndex`.

### Live Demo
https://jsbin.com/dawihibuva/1/edit?html,console,output

### Steps to reproduce

1. Visit the demo above
2. Request the indexes be initialized by clicking the 'Show Indexes' button
3. Update the size of the list by clicking the 'Resize List' button
4. Check the new value of the indexes by clicking the 'Show Indexes' button

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Safari 9
- [x] Safari 8
- [x] Safari 7
- [x] Edge
- [x] IE 11
- [x] IE 10

Westbrook:
Fixes #451 

`lastVisibleIndex` and `firstVisibleIndex` are both read or set variables, so they need to be nulled out in `_resizeHandler` when the list is resized or they will persist stale data until the list is scrolled or `items` are changed, which do the same.

### DIFF
--- a/iron-list.html
+++ b/iron-list.html
@@ -817,6 +817,10 @@ will only render 20.
           parseInt(styles['padding-top'], 10);
       this._isRTL = Boolean(styles.direction === 'rtl');
       this._viewportWidth = this.$.items.offsetWidth;
+      var viewportHeight = this._scrollTargetHeight;
+      if (viewportHeight !== this._viewportHeight) {
+        this._resetScrollPosition(this.scrollTarget.scrollTop);
+      }
       this._viewportHeight = this._scrollTargetHeight;
       this.grid && this._updateGridMetrics();
     },
@@ -1426,7 +1430,6 @@ will only render 20.
       this._updateScrollerSize(true);
       this._positionItems();
       this._resetScrollPosition(this._physicalTop + this._scrollOffset + targetOffsetTop);
-
       this._increasePoolIfNeeded(0);
       // clear cached visible index.
       this._firstVisibleIndexVal = null;
--- a/iron-list.html
+++ b/iron-list.html
@@ -1449,10 +1449,10 @@ will only render 20.
      * when the element is resized.
      */
     _resizeHandler: function() {
-      // clear cached visible index.
-      this._firstVisibleIndexVal = null;
-      this._lastVisibleIndexVal = null;
       this._debounceRender(function() {
+        // clear cached visible index.
+        this._firstVisibleIndexVal = null;
+        this._lastVisibleIndexVal = null;
         // Skip the resize event on touch devices when the address bar slides up.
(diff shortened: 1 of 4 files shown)


iron-list.html
```diff
@@ -817,6 +817,10 @@
           parseInt(styles['padding-top'], 10);
       this._isRTL = Boolean(styles.direction === 'rtl');
       this._viewportWidth = this.$.items.offsetWidth;
+      var viewportHeight = this._scrollTargetHeight;
+      if (viewportHeight !== this._viewportHeight) {
+        this._resetScrollPosition(this.scrollTarget.scrollTop);
+      }
```
Westbrook: These lines could be on the chopping block for this PR, they are a little tricky. With the additional `requestAnimationFrame`s in the unit tests then this code isn't needed. However, I'm not sure if that approach more accurately recreates real life usage in the unit test context, or if acts as a work around to this code being absent.

Originally, I meant to submit this PR without this code, but with the phantom test failures I put it back in to see if I could get them to pass. With the addition of `list.style.width = '320px';` below, removing this could go either way. Thoughts?

test/grid.html
```diff
@@ -241,7 +241,57 @@
         });
       });
     });
- 
+
+    test('last visible items should change on resize', function(done) {
+      flush(function() {
+        // Physical scroll bar causes there to be less columns than expected.
+        // Force list to be wide enough to allow for 3 columns and the scroller.
+        list.style.width = '320px';
```
Westbrook: This is a bit of a hack, but I wasn't sure how else to handle the issue. The initial layout assumes full width, but the later layout in response to the `iron-resize` uses full width minus one physical scroll bar, which breaks the anticipated layout. Not sure if I should do some deeper changes to `x-grid`, create another helper, just go with this approach being not a lot of tests use `iron-resize`, or whether this merits its own issue where the layout math is updated in `iron-list`.

iron-list.html
keanulee: I prefer that this be removed in favor of keeping the width setter in the test instead.

PR #454, opened 2017-06-18
Update layout correctly when changing the value of `grid`

### LINKED ISSUE #453: Toggling `grid` doesn't correctly update the layout.
### Description
Toggling `grid` doesn't correctly update the layout.

### Expected outcome
Switching `grid` to `true` should cause columns to be formed and switching to `false` should remove those columns, and allow the end of the list to be viewable.

### Actual outcome
Layout is updated erroneously, new `translate3d` is not applied.

### Live Demo
https://jsbin.com/yarevisawu/1/edit?html,console,output

### Steps to reproduce

1. Visit the demo URL above.
2. Click "Toggle Grid".
3. See the erroneous layout.
4. Scroll around a bit to get the layout updated, notice you can't scroll all the way to the last items.
5. Click "Toggle Grid".
6. See the erroneous layout.

### Browsers Affected

- [x] Chrome
- [x] Firefox
- [x] Safari 9
- [x] Safari 8
- [x] Safari 7
- [x] Edge
- [x] IE 11
- [x] IE 10

Westbrook:
Depends on #452 
Fixes #453 

When grid is changed a number of internal metrics need to be updated and the layout needs to be recast so that the list displays as expected.

### DIFF
--- a/iron-list.html
+++ b/iron-list.html
@@ -1459,10 +1459,10 @@ will only render 20.
      * when the element is resized.
      */
     _resizeHandler: function() {
-      // clear cached visible index.
-      this._firstVisibleIndexVal = null;
-      this._lastVisibleIndexVal = null;
       this._debounceRender(function() {
+        // clear cached visible index.
+        this._firstVisibleIndexVal = null;
+        this._lastVisibleIndexVal = null;
         // Skip the resize event on touch devices when the address bar slides up.
         var delta = Math.abs(this._viewportHeight - this._scrollTargetHeight);
         this.updateViewportBoundaries();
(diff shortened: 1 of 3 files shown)


iron-list.html
```diff
@@ -1058,6 +1063,13 @@
       this.templatize(this._userTemplate, this.mutableData);
     },
 
+    _gridChanged: function(newGrid, oldGrid) {
+      if (typeof oldGrid === 'undefined') return;
+      this.fire('iron-resize');
```
Westbrook: Not sure if `.fire` is either correct when also supporting the 2.x ecosystem or worth the overhead when I can  just call `_resizeHandler` directly. Thoughts?

keanulee: `this.notifyResize()` (https://github.com/PolymerElements/iron-resizable-behavior/blob/master/iron-resizable-behavior.html#L81) is what you're looking for.

PR #455, opened 2017-06-18
Make sure that physical items are always managed as complete rows.

### LINKED ISSUE #446: Grid view iron-list displaying items in incorrect position
### Description
Iron-list in grid view sometimes shows list items in incorrect position on the grid. This results in double-ups, items showing out of order and switching behaviour when scrolling. In my particular app this gives me issues as my items need to be sorted in a specific way.

### Expected outcome

Items should appear in grid in the same order that they appear in in the items array

### Actual outcome

Some items appear in grid in different order

### Live Demo
http://jsbin.com/gipovanifu/edit?html,output

### Steps to reproduce
This can be a little tricky to spot in the jsbin but basically keep your eyes on the first column. As each row has 5 items, only items with indices ending in 0 or 5 should appear in this column.

Basically scroll around a bit aggressively until you see any index that isn't a multiple of 5 in the first column. Should look like the screenshot below which I took from the jsbin.

screen shot 2017-06-05 at 16 21 16

### Browsers Affected
I've been testing on chrome but I've seen it happen on firefox and iOS/desktop safari so far too.

- [ ] Chrome
- [ ] Firefox
- [ ] Safari 9
- [ ] Safari 8
- [ ] Safari 7
- [ ] Edge
- [ ] IE 11
- [ ] IE 10

Westbrook:
Fixes #446

When grid items are represented by physical items that don't add up to complete rows it is possible for the recycler to skip items causing them to not be shown and other items to be positioned in the wrong column.

### DIFF
--- a/iron-list.html
+++ b/iron-list.html
@@ -976,7 +976,7 @@ will only render 20.
           DEFAULT_PHYSICAL_COUNT, this._virtualCount - this._virtualStart);
       nextPhysicalCount = this._convertIndexToCompleteRow(nextPhysicalCount);
       var correction = nextPhysicalCount % this._itemsPerRow;
-      if (correction && nextPhysicalCount - correction <= this._physicalCount) {
+      if (correction && nextPhysicalCount - correction < this._physicalCount) {
         nextPhysicalCount += this._itemsPerRow;
       }
       nextPhysicalCount -= correction;
--- a/iron-list.html
+++ b/iron-list.html
@@ -975,11 +975,13 @@ will only render 20.
       var nextPhysicalCount = this._clamp(this._physicalCount + count,
           DEFAULT_PHYSICAL_COUNT, this._virtualCount - this._virtualStart);
       nextPhysicalCount = this._convertIndexToCompleteRow(nextPhysicalCount);
-      var correction = nextPhysicalCount % this._itemsPerRow;
-      if (correction && nextPhysicalCount - correction <= this._physicalCount) {
-        nextPhysicalCount += this._itemsPerRow;
+      if (this.grid) {
+        var correction = nextPhysicalCount % this._itemsPerRow;
+        if (correction && nextPhysicalCount - correction <= this._physicalCount) {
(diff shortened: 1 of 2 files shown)


Westbrook:
Also, I'm interested in the following TODO:
```
// TODO: investigate why this is flaky.
// assert.equal((list._physicalEnd + 1) % list._itemsPerRow, 0, '_physicalEnd');
```
What sort of flakiness had been exhibited by this test? I'm unable to get the test to fail against `master` when run locally to confirm, but I'd guess that the change herein would have an effect on what you were seeing.

iron-list.html
```diff
@@ -1668,6 +1677,16 @@
...
     },
 
+    /**
+     * Converts a random index to the index of the item that completes it's row.
+     * Allows for better order and fill computation when grid == true.
+     */
+    _convertIndexToCompleteRow: function(idx) {
+      // when grid == false _itemPerRow can be unset.
+      this._itemsPerRow = this._itemsPerRow || 1;
+      return Math.ceil(idx / this._itemsPerRow) * this._itemsPerRow;
```
keanulee: Avoid multiply/divide when grid is false:

```js
return this.grid ? ... : idx;
```

iron-list.html
```diff
@@ -973,6 +974,12 @@
     _increasePoolIfNeeded: function(count) {
       var nextPhysicalCount = this._clamp(this._physicalCount + count,
           DEFAULT_PHYSICAL_COUNT, this._virtualCount - this._virtualStart);
+      nextPhysicalCount = this._convertIndexToCompleteRow(nextPhysicalCount);
+      var correction = nextPhysicalCount % this._itemsPerRow;
+      if (correction && nextPhysicalCount - correction <= this._physicalCount) {
+        nextPhysicalCount += this._itemsPerRow;
+      }
+      nextPhysicalCount -= correction;
```
keanulee: Only do this work for grid: `if (this.grid) ...`

PR #457, opened 2017-06-23
Select items

### LINKED ISSUE #436: Select processing of large groups of selects takes too long

Westbrook:
Fixes #436 

Replies on:
Polymer/polymer#4699
Polymer/polymer#4705

Adds code paths and tests to support selecting all of the items in the list or batch selecting a subset of the list. Further the ideas introduces in #437, this relies on extensions to both Polymer 1.x and 2.x to empower `array-selector` with selectItems and selectAll functionality and forks the application of that code as needed in this _hybrid_ component.

### DIFF
--- a/iron-list.html
+++ b/iron-list.html
@@ -1541,7 +1541,7 @@ will only render 20.
       if (arguments.length > 1 || (typeof items !== 'undefined' && !Array.isArray(items))) {
         items = !!Array.from ? Array.from(arguments) : [].slice.call(arguments);
       }
-      if (!indexes.length) {
+      if (typeof indexes === 'undefined' || !indexes.length) {
         items = this.items;
       } else {
         item = indexes.map(function(idx) {
--- a/iron-list.html
+++ b/iron-list.html
@@ -1531,6 +1531,7 @@ will only render 20.
 
     // TODO: finish me!!
     selectIndexesV1: function(indexes) {
+      var items = [];
       var item = {};
       var model = {};
       var skey = '';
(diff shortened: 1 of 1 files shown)


Westbrook:
Based on the response from the Polymer team in https://github.com/Polymer/polymer/pull/4699#issuecomment-331311907 would a PR that included a custom `iron-list-selector` element be something that would be acceptable? Trying to see whether I should push back on that comment or whether that seems reasonable that this level of responsibility be passed down to the `iron-list` component.

I guess one alternate possibility would be to have a stand alone component/repo that handled these responsibilities and then require it in `iron-list` but I’m not sure the right place to expect that to be managed. Thoughts?

Westbrook:
Would that still block this PR on 1.x not having this capability? Can we move to a non-hybrid `iron-list`, yet? Just trying to figure out the right way to move this forward. Thanks for any suggestions.

keanulee:
I would lean towards just baking `iron-list-selector` functionality right into `iron-list` if that's possible. Having different code paths to maintain hybrid compatibility would be acceptable.

PR #458, opened 2017-06-23
Fix _gridChanged timing for IE/Edge

keanulee:


iron-list.html
```diff
@@ -1072,9 +1072,9 @@
 
     _gridChanged: function(newGrid, oldGrid) {
       if (typeof oldGrid === 'undefined') return;
+      newGrid && this._updateGridMetrics();
       this.notifyResize();
       Polymer.flush ? Polymer.flush() : Polymer.dom.flush();
-      this._updateGridMetrics();
```
frankiefu: The `newGrid` check looks fine. But just by looking at the code it seems like `updateGridMetrics` should be called after `flush`, to make sure all physical items are distributed?

PR #461, opened 2017-07-01
Make iron-list.html ES5 compatible

balloob:
`iron-list.html` had 1 piece of code using a syntax introduced in ES2015, and thus not compatible with ES5. As not everyone consuming the hybrid elements has an ES2015 conversion pipeline setup, it would be great if elements could stick to ES5 syntax.

If this is not a goal, feel free to close this PR.

### DIFF
--- a/iron-list.html
+++ b/iron-list.html
@@ -1882,7 +1882,7 @@ will only render 20.
       }
     },
 
-    _forwardProperty(inst, name, value) {
+    _forwardProperty: function(inst, name, value) {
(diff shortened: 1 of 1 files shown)


balloob:
I signed it!

balloob:
Could you create a new release @keanulee ?

keanulee:
@balloob Done https://github.com/PolymerElements/iron-list/releases/tag/v2.0.4

PR #472, opened 2017-07-20
Fix forwardItemPath function

### LINKED ISSUE #468: cannot read property 'offsetHeight' of undefined
The code here :

https://github.com/PolymerElements/iron-list/blob/master/iron-list.html#L1263

```
        oldPhysicalSize += this._physicalSizes[pidx] || 0;
        this._physicalSizes[pidx] = this._physicalItems[pidx].offsetHeight;
        newPhysicalSize += this._physicalSizes[pidx];
        this._physicalAverageCount += this._physicalSizes[pidx] ? 1 : 0;
      }, itemSet);
```

the assignment to `oldPhysicalSize` checks if `this._physicalSizes[pidx]` is defined before assignment, yet the following line assumes the corresponding `offsetHeight` property does exist. If it doesn't exist, as in my case, this results in the error:

```
cannot read property 'offsetHeight' of undefined
```

Examining _physicalSizes and _physicalItems, it seems like they contain only a small subset of my app's `items`...indexed 0..22. However, the value of `pidx` is `4390`. As a matter of interest, `vidx`, which is never used in that function, has value `8737`). Note that iron-list's `items` list is `4391`, so I'm curious where `8737` comes from.

The stack trace when I block on this line shows that it originates from a `this.set` on a property of one of the items in `items`.

If anyone has any ideas how I can resolve this problem, please let me know.

AlbertoFdzM:
Fixes #468

### DIFF
--- a/iron-list.html
+++ b/iron-list.html
@@ -1122,9 +1122,11 @@ will only render 20.
       var offscreenItem = this._offscreenFocusedItem;
       var isIndexRendered = this._isIndexRendered(idx);
       var inst;
+      var pidx;
 
       if (isIndexRendered) {
-        inst = this.modelForElement(this._physicalItems[this._getPhysicalIndex(idx)]);
+        pidx = this._getPhysicalIndex(idx);
+        inst = this.modelForElement(this._physicalItems[pidx]);
       } else if (offscreenItem) {
         inst = this.modelForElement(offscreenItem);
       }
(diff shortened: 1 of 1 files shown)


keanulee:
Thanks @AlbertoFdzM

AlbertoFdzM:
Thanks for adding the test

PR #478, opened 2017-07-27
Ensure different debouncers don't cancel each other

### LINKED ISSUE #475: _itemsChanged(items.*) does not re-render list implicitly.
Referring to: https://github.com/PolymerElements/iron-list/blob/368f75fae81cd7696b710635eccacbca7a6a93b1/iron-list.html#L1067

### Description

Changes to the `items` array does not cause the iron-list re-render. A manual call to `Polymer.flush()` is still required.

### Expected outcome

The list is automatically rerendered after items has been changed with a completely new array.

### Actual outcome

The call here https://github.com/PolymerElements/iron-list/blob/368f75fae81cd7696b710635eccacbca7a6a93b1/iron-list.html#L1085 does not cause a Polymer.flush(), and indeed it never happens until the user scrolls more. 

### Steps to reproduce
1. Load a list.
2. Reload the list by replacing the `items` array completely.
3. The list does not get re-rendered.

### Browsers Affected

- [x] Chrome
- [ ] Firefox
- [ ] Safari 9
- [ ] Safari 8
- [ ] Safari 7
- [ ] Edge
- [ ] IE 11
- [ ] IE 10

keanulee:
Fixes #475

### DIFF
--- a/iron-list.html
+++ b/iron-list.html
@@ -766,7 +766,7 @@ will only render 20.
     },
 
     attached: function() {
-      this._debounceRender(this._render, ANIMATION_FRAME);
+      this._debounce('_render', this._render, ANIMATION_FRAME);
       // `iron-resize` is fired when the list is attached if the event is added
       // before attached causing unnecessary work.
       this.listen(this, 'iron-resize', '_resizeHandler');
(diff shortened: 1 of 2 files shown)


keanulee:
The `1.x` branch is no longer maintained. However, iron-list 2.x is compatible with Polymer 1.9 and later (https://github.com/PolymerElements/iron-list/blob/master/bower.json#L23).

kito99:
OK, thanks @keanulee .

PR #486, opened 2017-09-12
Fix #485

### LINKED ISSUE #485: _isVisible should check size of scrollTarget
[The `_isVisible` method](https://github.com/PolymerElements/iron-list/blob/master/iron-list.html#L689-L691) checks `offsetWidth` and `offsetHeight` of the `iron-list` element to determine if it the list is visible or not. The method is used to determine whether the list should be rendered and to trigger adding/removing the scroll listener, I'm assuming as a performance optimization.

The problem with always checking the `iron-list` element itself is that if an external `scrollTarget` is used, the size if the list is not a good indicator for whether it is 'visible'. For example, if the list is empty and not sized explicitly, the `iron-list` elements height will be `0`, which will prevent a render even if the corresponding `scrollTarget` is visible. I've also had problems with the scroll handler being removed and not added back if the `scrollTarget` was hidden and shown while the list itself was empty.

The obvious and simple fix is to check the size of the `scrollTarget` instead of the `iron-list` element itself. Gonna submit a PR for this soon.

MaKleSoft:
Check size of scroll target instead of iron-list element to determine whether list is 'visible'. Fixes #485

### DIFF
--- a/iron-list.html
+++ b/iron-list.html
@@ -687,7 +687,7 @@ will only render 20.
     * True if the current list is visible.
     */
     get _isVisible() {
-      return Boolean(this.offsetWidth || this.offsetHeight);
+      return Boolean(this.scrollTarget.offsetWidth || this.scrollTarget.offsetHeight);
     },
 
     /**


MaKleSoft:
I'm guessing this won't be merge now 😄. Closing

PR #487, opened 2017-09-13
More performant attempt at solving #300

admwx7:
@blasten how does this hold up instead?  Figured if we leave it up to the implementor then the performance hit is on them, in my specific case I actually only need 2 properties to update so just calling notifyPath on those two is the only hit I actually need, just finished testing (moving to hybrid for my element) and it seems to work perfectly fine, at least for user interactions, having some weird issues with programatic updates but that's something I did, not iron-list.


admwx7:
@keanulee figured I should ping you, believe you conceded to now having ownership of iron-list to me during the summit.

PR #490, opened 2017-10-07
Proper annotations

valdrinkoshi:
Makes closure compiler happy!


valdrinkoshi:
@keanulee gentle ping. This PR makes closure compiler happy \o/
Not sure why tests fail, ideally I should have set the commit message with `[skip ci] ...` (as it only modifies comments) but now is too late..

PR #491, opened 2017-10-12
Ensure _physicalSizes contains numbers

valdrinkoshi:
I've experimented few race conditions where having `this._physicalSizes[i] === undefined` would end up making properties like `this._physicalTop = NaN`.

This PR aims to clear these cases by ensuring the array is always containing numbers, hence ~uses [Array.fill](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/fill)~ (lol, just read the browser support, was mislead by not seeing any red color in the table) fills the array with 0s when the pool is increased.

Not really sure how to reproduce the race condition, open to suggestions.

### DIFF
--- a/iron-list.html
+++ b/iron-list.html
@@ -976,7 +976,7 @@ will only render 20.
         var ts = window.performance.now();
         // Concat arrays in place.
         [].push.apply(this._physicalItems, this._createPool(delta));
-        [].push.apply(this._physicalSizes, new Array(delta));
+        [].push.apply(this._physicalSizes, new Array(delta).fill(0));
         this._physicalCount = this._physicalCount + delta;
         // Update the physical start if it needs to preserve the model of the focused item.
         // In this situation, the focused item is currently rendered and its model would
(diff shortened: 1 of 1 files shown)


iron-list.html
```diff
@@ -976,7 +976,7 @@
         var ts = window.performance.now();
         // Concat arrays in place.
         [].push.apply(this._physicalItems, this._createPool(delta));
-        [].push.apply(this._physicalSizes, new Array(delta));
+        [].push.apply(this._physicalSizes, new Array(delta).fill(0));
```
keanulee: So IE doesn't support `.fill()`, hence the test failures. LGTM otherwise.

valdrinkoshi: lol, yeah I was mislead by not seeing any red color in the MDN support table :p
Fixed in https://github.com/PolymerElements/iron-list/pull/491/commits/3772d0937fde34d2169335188bb5306b0fb82c76

PR #497, opened 2017-11-16
Handle addedCount of splice [Fix #496]

### LINKED ISSUE #496: Handling of items splice does not take splice length into account
### Description
The logic in `_itemsChanged` looks at the starting index of a splice when determining whether `_render` should be triggered, but it doesn't consider the length of the splice. This means that the view is not updated if the splice starts before the currently visible range of items, but ends in the range or after it.

### Live Demo
https://jsbin.com/fifijovute/edit?html,output

### Steps to reproduce

1. Click Scroll
2. Click Splice
3. Expected result is that items 495 - 499 have been updated, actual result is that those items are not updated until scrolling away from the affected area and back again.

If clicking Splice before Scroll, everything works because the starting index of the splice is then in  view when the splice happens.

keanulee:


frankiefu:
LGTM

PR #508, opened 2018-02-14
Fix strict modules

e111077:
1. Added 'use strict' to tests
2. declared `item` in `helpers.html`
3. made functions visible to tests via window


e111077:
Failing tests seem to be the same ones on IE as master, but I don't think they're related to my PR

PR #547, opened 2018-11-09
Undo some modulizer funkyness

rictic:


e111077:
Things are failing

rictic:
Huh, were they passing at master?

keanulee:
Maybe `Polymer` identifier is clashing with `Polymer.Collection` type? https://github.com/PolymerElements/iron-list/blob/master/iron-list.js#L493

keanulee:
`Polymer.Collection` is no longer a thing as of Polymer 2, so `this._collection` and related code paths could be removed.

keanulee:
Nope, removing references to `Polymer.Collection` didn't work for me locally. Something with not importing `Polymer` as some other identifier (e.g. `Polymer$0`).

e111077:
I think that the Polymer namespace may be conflicting with the Typescriptified Polymer namespace. I believe that typescript may be looking into polymer-fn for these types instead of the externs

PR #562, opened 2019-10-10
[1.x] Fix issue where _increasePoolIfNeeded could loop indefinitely. Fixes #559

### LINKED ISSUE #559: Physical items can grow unbounded when scrolling heterogeneously-sized list
### Description
`iron-list` can suffer from a runaway condition where `_increasePoolIfNeeded` can loop indefinitely due to the physical average becoming large enough (due to heterogeneously-sized items) to position the physical top lower than the scroll position, causing `_isClientFull` to return true.

### Expected outcome

Count of physical items never grows significantly larger than needed to cover the viewport.

### Actual outcome

Count of physical items grows close to the virtual size of a large virtual list.

### Live Demo
https://glitch.com/edit/#!/circular-breeze?path=index.html

### Steps to reproduce
1. Create list with large `items` array and heterogeneously-sized items
2. Scroll to a point in the list such that the physical average becomes larger than at the start of the list
3. Scroll back to the top of the list; if the physical average is significantly large enough, `_scrollHandler` will reset the `_physicalTop` to a value larger than the `_scrollPosition`.
4. Fire an `iron-resize` event at the list. This will cause `_increasePoolIfNeeded` to check `_isClientFull`, which will return falsey because the items don't cover the viewport, and will loop indefinitely.

This bug appears to have been at least partially introduced in https://github.com/PolymerElements/iron-list/pull/435/files, where the `_physicalTop` is re-calculated purely as a function of `physicalAverage`, losing a guarantee that the `_physicalTop` is no longer above the start of the scroll position.

This could also explain some reported issues in https://github.com/PolymerElements/iron-list/issues/506 (although the OP of that issue seems to have been a user error with list sizing, others report similar runaway physical item creation behavior with a correctly sized list). I created a separate issue here since a GIF posted in that issue appears to have uniformly-sized items, and I _believe_ this particular bug only occurs when the `_physicalAverage` used for estimation random-access scrolling becomes significantly larger than the physical items in the viewport.

### Browsers Affected

- [X] Chrome (only confirmed on Chrome; didn't test other browsers)
- [ ] Firefox
- [ ] Safari 9
- [ ] Safari 8
- [ ] Safari 7
- [ ] Edge
- [ ] IE 11
- [ ] IE 10

kevinpschaaf:
Ensure the _physicalTop is never recalculated to be below the current scroll position, and make sure cached scroll positions are used in methods which may be called outside of a scroll event handler for internal consistency.


iron-list.html
```diff
@@ -882,7 +885,8 @@
       var physicalCount = this._physicalCount;
       var top = this._physicalTop + this._scrollOffset;
       var bottom = this._physicalBottom + this._scrollOffset;
-      var scrollTop = this._scrollTop;
+      // This may be called outside of a scrollHandler, so use last cached position
+      var scrollTop = this._scrollPosition;
```
sorvell: Found the comment confusing. Isn't `_scrollPosition` the more "live" of the two properties (`_scrollPosition` and `_scrollTop`)?

sorvell: Actually `_scrollTop` is "live." It comes from `iron-scroll-target-behavior`.

iron-list.html
```diff
@@ -846,8 +846,11 @@
         var idxAdjustment = Math.round(delta / this._physicalAverage) * this._itemsPerRow;
         this._virtualStart = this._virtualStart + idxAdjustment;
         this._physicalStart = this._physicalStart + idxAdjustment;
-        // Estimate new physical offset.
-        this._physicalTop = Math.floor(this._virtualStart / this._itemsPerRow) * this._physicalAverage;
+        // Estimate new physical offset based on the virtual start (but never allow
+        // to be more than the current scrollTop)
+        this._physicalTop = Math.min(
```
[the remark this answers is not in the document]
kevinpschaaf: Improved comment.  Opted not to constrain other bound since it was non-trivial to repro.

iron-list.html
kevinpschaaf: Added comment to declaration of `_scrollPosition` that it is a cached version of `_scrollTop`

test/basic.html
```diff
@@ -290,15 +297,45 @@
...
+    // This test catches a runaway condition where _increasePoolIfNeeded
+    // could recurse indefinitely due to the physical average becoming
+    // large enough (due to heterogeneously-sized items) to position the
+    // physical top lower than the scroll position, causing _isClientFull
+    // to return true; this 
+    test('Issue #559: scroll heterogeneously sized list', function(done) {
+      list.scrollTop = 2300;
+      requestAnimationFrame(function() {
+        list.scrollTop = 0;
+        list.fire('iron-resize');
```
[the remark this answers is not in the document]
kevinpschaaf: Removed, not needed in P1 version because `_increasePoolIfNeeded` is called on every scroll (in P2 it's not called on a random access).

